PR #55, opened 2020-09-07
Implements #45 - From/to bytes for PoseidonCipher

vlopes11:


CPerezz (approved):
LGTM. Can you bump the patch version please?

PR #67, opened 2020-09-30
PoseidonTree - Generic annotation

vlopes11:
We should have a generic annotation so it can be extended to specific types.

One use case is the blind bid where we need to annotate block height ranges.


CPerezz (approved):
LGTM

PR #68, opened 2020-10-01
Poseidon Tree with generic annotation

vlopes11:


CPerezz (approved):
LGTM. But this is a minor version at least for me. Shouldn't we bump to v0.8.1. You're not breaking the API.

PR #91, opened 2020-11-04
PoseidonLeaf pos API consistency

vlopes11:
### Changed
- PoseidonLeaf pos setter for API consistency with Phoenix

### Removed
- PoseidonCipher std::io implementations

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,12 +107,12 @@ impl PoseidonLeaf<MemStore> for DataLeaf {
     }
 
     // Position on the tree
-    fn tree_pos(&self) -> u64 {
+    fn pos(&self) -> u64 {
         self.pos
     }
 
     // Method used to set the position on the tree after the `PoseidonTree::push` call
-    fn tree_pos_set(&mut self, pos: u64) {
+    fn set_pos(&mut self, pos: u64) {
         self.pos = pos;
     }
 }
--- a/README.md
+++ b/README.md
@@ -112,8 +112,8 @@ impl PoseidonLeaf<MemStore> for DataLeaf {
     }
 
     // Method used to set the position on the tree after the `PoseidonTree::push` call
-    fn tree_pos_mut(&mut self) -> &mut u64 {
-        &mut self.pos
+    fn tree_pos_set(&mut self, pos: u64) {
+        self.pos = pos;
     }
 }
 
--- a/src/cipher/cipher.rs
+++ b/src/cipher/cipher.rs
@@ -12,11 +12,7 @@ use dusk_plonk::jubjub::AffinePoint;
 use dusk_plonk::prelude::*;
 use hades252::{ScalarStrategy, Strategy, WIDTH};
 
-use super::{
-    CIPHER_BYTES_SIZE, CIPHER_SIZE, ENCRYPTED_DATA_SIZE, MESSAGE_CAPACITY,
-};
-
-use std::io;
+use super::{CIPHER_BYTES_SIZE, CIPHER_SIZE, MESSAGE_CAPACITY};
 
 pub use super::CipherError;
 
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -112,8 +112,8 @@ impl PoseidonLeaf<MemStore> for DataLeaf {
     }
 
     // Method used to set the position on the tree after the `PoseidonTree::push` call
-    fn tree_pos_mut(&mut self) -> &mut u64 {
-        &mut self.pos
+    fn tree_pos_set(&mut self, pos: u64) {
```
ZER0: Is the `tree` prefix mandatory? Couldn't the trait just needs to implement `pos()` (the getter) and `set_pos()` (the setter) as phoenix's note does?

If you think it's not clear enough is ok, I'll leave this up to you, but the naming in that case should be `set_tree_pos`.

CPerezz (approved):
LGTM

src/cipher/mod.rs
```diff
@@ -16,9 +16,6 @@ pub const CIPHER_SIZE: usize = MESSAGE_CAPACITY + 1;
 /// Number of bytes used by from/to bytes `PoseidonCipher` function
 pub const CIPHER_BYTES_SIZE: usize = CIPHER_SIZE * 32;
 
-/// Bytes consumed on serialization of the poseidon cipher
-pub const ENCRYPTED_DATA_SIZE: usize = CIPHER_SIZE * 32;
```
ZER0, line 20: Don't remove this. I believe is still used across other repos since it was a Public API. You can add a TODO note in order to unify with the `CIPHER_BYTES_SIZE` or something like that. I believe that `ENCRYPTED_DATA_SIZE` is used more across our repos than `CIPHER_BYTES_SIZE`, also.

src/cipher/cipher.rs
```diff
@@ -251,45 +242,3 @@ impl PoseidonCipher {
         [domain, length, ks0, ks1, nonce]
     }
 }
-
-impl io::Write for PoseidonCipher {
```
ZER0, line 255: I thought it was used for the from / to bytes impl, but good to know we can safely remove it without changing the implementation of the methods above!

ZER0 (changes requested):
LGTM, but there are few changes need to be done.

README.md
vlopes11: We have (imo) a limitation in Rust that will not grant namespacing for trait implementations. So if we have a very generic name for a trait that should be very generic, as is the `PoseidonLeaf`, then if the implementor needs to implement a method with the same name, the references to the method will be very ugly such as `<SomeType as PoseidonLeaf>::set_pos(&mut leaf, pos)`

The `tree_` prefix is a sort of workaround for the lack of a namespace, but I'm honestly not happy with this solution.

I could rename them both to `pos` and `set_pos`; what do you think?

ZER0: It's basically the issue we had with the `read` impl but I'm fine with that because it's only happened because the macro. We can ensure we're using the proper trait just doing `PoseidonLeaf::set_pos(&mut leaf, pos)` whenever arise the issue – it means, both of the trait are in the same scope.

I'm fine then with just use `pos` and `set_pos`, then.

vlopes11: Done

src/cipher/mod.rs
vlopes11: Done

src/cipher/cipher.rs
vlopes11: Yep, it was meant to be an extension and never a requirement

src/cipher/cipher.rs
```diff
@@ -147,11 +143,6 @@ impl PoseidonCipher {
         MESSAGE_CAPACITY
     }
 
-    /// Bytes consumed on serialization of the poseidon cipher
-    pub fn serialized_size() -> usize {
-        ENCRYPTED_DATA_SIZE
-    }
```
ZER0: This is also should be kept for the same reason of `ECRYPTED_DATA_SIZE`, sorry I missed the last time. I'm not sure if it's widely used, but I remember it is used as standard function for a lot of other types.

Also, I would suggest to convert to to a `const fn` (as we have for the other types).

ZER0 (changes requested):

src/cipher/cipher.rs
vlopes11: Done

ZER0 (approved):
Thanks!

PR #103, opened 2020-12-04
PoseidonMaxAnnotation borrow fix latest microkelvin

vlopes11:
Last version of microkelvin forces us to use `Borrow<Max<u64>>` instead of `Borrow<u64>` for the max key


CPerezz (approved):
Lgtm!

ZER0 (approved):

PR #108, opened 2021-01-26
Adapt Poseidon252 with library that uses dusk-bytes

ZER0:
- Remove `hex` crate for unit test in favor of `dusk-bytes`


vlopes11 (approved):
LGTM!

PR #123, opened 2021-03-30
Merkle opening constant circuit description

### LINKED ISSUE #122: Merkle Opening should use fixed witness positions
The merkle opening updates the index of the witness according to the provided offset.

This may lead to problems with the verifier key because it may cause the circuit to not be constant in some specific cases.

One example of a problem is described here:
https://github.com/dusk-network/plonk-vk

vlopes11:
PLONK requires the circuit description + witness indexes to be constant
so the same verifier data can be reused to check different proofs.

The previous implementation contained an optimization that swaps the
witness index to reduce the number of gates.

This optimization may lead to inconsistencies when reusing verifier keys
because the verification will depend on the disposition of the witness
data, and this is undesirable.

Resolves #122


CPerezz (approved):
LGTM. Could we do a pre-release just in case we need tochange anything else?

PR #163, opened 2021-08-12
Change merkle opening to constrain leaf

### LINKED ISSUE #162: Tree: Merkle opening should constrain leaf
Currently the merkle opening is not constraining the leaf.

This forces the consumer to reproduce the same logic of the merkle bits checker that is internal to poseidon.

The merkle_opening gadget should take the leaf as witness and constrain it to the first level of the opening branch.

vlopes11:
If the merkle opening doesn't constrain the leaf, the consumer of the
library need to be aware of the poseidon internals, select the merkle
bit and constrain it manually.

By constraining it in the opening gadget, we consume only a single gate
and protect the user from poseidon internals.

Resolves #162

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Change merkle opening to constrain leaf [#162]
+
 ## [0.22.0] - 2021-07-27
 
 ### Changed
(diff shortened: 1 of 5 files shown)


CPerezz (approved):
Feels correct. Idk when that changed. But definitely it is somehting that I don't like at all. When did this change to become something on which the leaf is not passed to the merkle-proof fn?? It's nonsensical no??

LGTM once checks pass

ZER0 (approved):

PR #168, opened 2021-12-17
Update sponge to export truncated hash

### LINKED ISSUE #167: Export `sponge::truncated_hash` regardless of `alloc` feature
Alloc shield plonk proving system. However, some of the functionality of the crate can be exposed without using plonk - including the truncated hash.

Truncated hash is required to safely hash from bls to jubjub in a plonk circuit.

N/A

**Additional context**
N/A

vlopes11:
Resolves #167

### DIFF
--- a/src/sponge/truncated.rs
+++ b/src/sponge/truncated.rs
@@ -51,6 +51,7 @@ pub fn hash(messages: &[BlsScalar]) -> JubJubScalar {
 ///
 /// The returned value is the hashed witness data computed as a variable and truncated
 /// to fit inside of a [`JubJubScalar`].
+#[cfg(feature = "alloc")]
 pub fn gadget(composer: &mut TurboComposer, message: &[Witness]) -> Witness {
     let zero = TurboComposer::constant_zero();
     let h = sponge::gadget(composer, message);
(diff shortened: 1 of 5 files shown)


src/sponge/truncated.rs
```diff
@@ -8,6 +8,7 @@
 
 use crate::sponge;
 use dusk_bls12_381::BlsScalar;
+use dusk_jubjub::JubJubScalar;
```
moCello: Why do you specifically import `JubJubScalar` here? afaik it should be included in `dusk_plonk::prelude::*` which we need for other types anyway.

src/sponge/truncated.rs
```diff
@@ -8,6 +8,7 @@
...
 
 use dusk_plonk::prelude::*;
```
ureeves: Unused import

ZER0 (changes requested):
I don't think I agreed with the public API here, it also seems another PR was landed before this one that substantially change the API in the opposite way it was designed for.

Specifically here I'm referring to the `sponge.rs`  that re-export `truncated::hash` as `truncated_hash` where we intentionally refactored in the past `Poseidon252` to avoid such things; so it looks like we're doing a step back in that direction.

ZER0 (approved):

PR #182, opened 2022-09-19
Fix merkle opening circuit

### LINKED ISSUE #181: Fix merkle opening circuit
Fix the merkle opening circuit and add tests that catch this in the future

The circuit would be valid for an invalid path

N/A

**Additional context**
N/A

moCello:
Fix merkle opening circuit and add tests to catch this bug in the future.

Resolves: #181

### DIFF
--- a/src/tree/branch.rs
+++ b/src/tree/branch.rs
@@ -33,7 +33,7 @@ use rkyv::{Archive, Deserialize, Serialize};
     archive_attr(derive(CheckBytes))
 )]
 pub struct PoseidonLevel {
-    level: [BlsScalar; dusk_hades::WIDTH],
+    pub(crate) level: [BlsScalar; dusk_hades::WIDTH],
     index: u64,
 }
 
@@ -79,7 +79,7 @@ impl AsRef<[BlsScalar]> for PoseidonLevel {
     archive_attr(derive(CheckBytes))
 )]
 pub struct PoseidonBranch<const DEPTH: usize> {
-    path: Vec<PoseidonLevel>,
+    pub(crate) path: Vec<PoseidonLevel>,
 }
 
 impl<const DEPTH: usize> PoseidonBranch<DEPTH> {
--- a/src/tree/zk.rs
+++ b/src/tree/zk.rs
@@ -15,65 +15,227 @@ pub fn merkle_opening<const DEPTH: usize>(
     branch: &PoseidonBranch<DEPTH>,
     leaf: Witness,
 ) -> Witness {
-    let zero = TurboComposer::constant_zero();
-
-    let mut base = true;
-    let mut root = zero;
-
     // Generate a permutation container
-    let mut perm = [zero; dusk_hades::WIDTH];
+    let mut container = [TurboComposer::constant_zero(); dusk_hades::WIDTH];
 
-    // For every level, replace the level offset with needle,
-    // permute the level and set the needle to the next level
-    // to the poseidon result of the permutation
-    branch.as_ref().iter().for_each(|level| {
+    // Recalculate the root for the given branch
+    (0..DEPTH).fold(leaf, |root, depth| {
+        let level = branch.as_ref()[depth];
+
         // Create the bits representation of the offset as witness
+        // and make sure that offset points to a hash in the level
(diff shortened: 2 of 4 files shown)


src/tree/zk.rs
```diff
@@ -15,65 +15,237 @@ pub fn merkle_opening<const DEPTH: usize>(
...
-    // permutate the level and set the needle to the next level
-    // to the poseidon result of the permutation
-    branch.as_ref().iter().for_each(|level| {
         // Create the bits representation of the offset as witness
+        // and make sure that offset points to a hash in the level
         let offset_flag = level.offset_flag();
-
-        let mut sum = zero;
-        let mut offset_bits = [zero; dusk_hades::WIDTH - 1];
+        let mut sum = TurboComposer::constant_zero();
```
vlopes11, line 28: The original purpose of this block was to validate that we have only zeroes/ones, and that we have one, and only one `1`; the rest being `0`.

It was written before we had [component_boolean](https://docs.rs/dusk-plonk/0.12.0/dusk_plonk/constraint_system/struct.TurboComposer.html#method.component_boolean) available, which now completely supersede this implementation. What we need is to run the boolean component for every of these bits (so we guarantee they are either `0` or `1`, and check that they add to `1` (so we guarantee that we have one, and only one `1`).

src/tree/zk.rs
```diff
@@ -15,65 +15,237 @@ pub fn merkle_opening<const DEPTH: usize>(
...
-        level
-            .as_ref()
-            .iter()
-            .zip(perm.iter_mut())
-            .enumerate()
-            .for_each(|(i, (l, p))| {
-                *p = composer.append_witness(*l);
+        // Calculate the root for the next level
+        GadgetStrategy::gadget(composer, &mut container);
+        root = container[1];
```
vlopes11: Doesn't this look like a [fold](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fold) operation? Meaning: you initialize the `root` with `leaf`, mutate and return it for every iteration, and finally return the resulting `root`

src/tree/branch.rs
```diff
@@ -18,24 +18,25 @@ use nstack::NStack;
 /// Represents a level of a branch on a given depth
 #[derive(Debug, Default, Clone, Copy, Canon)]
 pub struct PoseidonLevel {
-    level: [BlsScalar; dusk_hades::WIDTH],
+    pub(crate) level: [BlsScalar; dusk_hades::WIDTH],
```
ureeves, line 36: Since the tests have been pulled to unit level, does this need to be `pub(crate)`?

ureeves (approved):
LGTM. One nit

src/tree/zk.rs
moCello: Good point! I'll address this in a new issue #188

src/tree/zk.rs
moCello: done, not sure if it generally becomes more readable with a `fold` but that is just my personal preference.

src/tree/branch.rs
moCello: Yes, otherwise the tests can not alter the levels to create a faulty branch.

src/tree/zk.rs
```diff
@@ -15,65 +15,227 @@ pub fn merkle_opening<const DEPTH: usize>(
...
+                // at the level offset
+                let constraint =
+                    Constraint::new().mult(1).a(bit).b(container[i]);
+                let expected = composer.gate_mul(constraint);
+                let constraint = Constraint::new().mult(1).a(bit).b(root);
+                let calculated = composer.gate_mul(constraint);
+
+                // Make sure that the hash at the offset of the current level
+                // matches the hash calculated in the previous level
+                composer.assert_equal(expected, calculated);
```
vlopes11, line 61: The `expected` here is a new witness, which can be anything. We need a way to assert that this new witness corresponds to the previous root

vlopes11 (changes requested):
One last question

vlopes11 (approved):
Never mind the previous review; I overlooked one of the constraints. lgtm!

PR #183, opened 2022-09-19
Change to use `ranno` annotations

### LINKED ISSUE #180: `PoseidonTree` is only available behind `canon`
**Describe the bug**
The `PoseidonTree` struct - one of the main features of this library, and a very important component for the genesis contracts - is only available behind the `canon` feature.
This means that to write a smart contract one is **required** to bring in the `canonical` library even though this is in principle not required.

**Expected behaviour**
The tree should be available without the `canon` feature, and ideally without bringing in `canonical`.

**Additional context**
This might be a larger problem than it seems. Since the trait allowing for tree traversal is `microkelvin::Compound`, which requires implementors to also be `Canon`, this might also require a refactor of `microkelvin` itself.

ureeves:
- Add `ranno` version `0.1` to dependencies
- Change `PoseidonTree` to build only with the `alloc` feature
- Change `PoseidonTree` to take a generic `Keyed` over the leaf type
  instead of a `PoseidonAnnotation`
- Make `PoseidonTree::new` const
- Update `microkelvin` from `0.15` to `0.17`
- Update `nstack` from `0.14.0-rc` to `0.16`
- Remove `canon` and `persistence` features
- Remove `Error` struct
- Remove `canonical` and `canonical-derive` from dependencies
- Remove `PoseidonMaxAnnotation`

Resolves #180


krl (approved):

Cargo.toml
```diff
@@ -1,59 +1,49 @@
...
-default = ["std", "canon"]
-alloc = ["dusk-hades/plonk"]
+default = ["std"]
 std = [
-    "dusk-hades/plonk",
-    "dusk-bls12_381/default",
     "dusk-jubjub/std",
+    "dusk-bls12_381/default",
+    "alloc",
 ]
```
moCello: Is there a specific reason why we keep the feature `std` around, when it is the same as `default`?

moCello (approved):
LGTM

PR #190, opened 2022-09-22
Make `PoseidonBranch` store the root separately

### LINKED ISSUE #189: Store the root of a branch seperately to the levels
Rethink the way we store the root in the `PoseidonBranch`.

At the moment the root of the `PoseidonBranch` is stored in the last level of the branch at the index 1. Ideally we want the root stored separately to the levels and not as an additional level.
This is also an issue due to the `rkyv` implementations. Since it is now a `Vec`, `PoseidonBranch<17>` would parse in the exact same way as `PoseidonBranch<16>` for example.

We could also want to store the levels not as a `Vec` but as an array instead.

**Additional context**
N/A

ureeves:
This also required a slight change in the algorithm computing the branch, to handle the root not being stored at the last position of the path.

Resolves #189

### DIFF
--- a/src/tree/branch.rs
+++ b/src/tree/branch.rs
@@ -8,10 +8,7 @@ use super::PoseidonLeaf;
 
 use crate::tree::PoseidonAnnotation;
 
-use alloc::vec::Vec;
-
 use core::borrow::Borrow;
-use core::iter;
 use core::ops::Deref;
 
 use dusk_bls12_381::BlsScalar;
@@ -79,7 +76,8 @@ impl AsRef<[BlsScalar]> for PoseidonLevel {
     archive_attr(derive(CheckBytes))
 )]
 pub struct PoseidonBranch<const DEPTH: usize> {
-    pub(crate) path: Vec<PoseidonLevel>,
+    pub(crate) path: [PoseidonLevel; DEPTH],
+    root: BlsScalar,
 }
 
 impl<const DEPTH: usize> PoseidonBranch<DEPTH> {
@@ -88,10 +86,7 @@ impl<const DEPTH: usize> PoseidonBranch<DEPTH> {
 
     /// Represents the root for a given path of an opening over a subtree
     pub fn root(&self) -> &BlsScalar {
-        self.path
-            .last()
-            .map(|l| l.deref())
-            .unwrap_or(&Self::NULL_ROOT)
+        &self.root
     }
 }
 
@@ -100,16 +95,6 @@ impl<const DEPTH: usize> Deref for PoseidonBranch<DEPTH> {
 
     fn deref(&self) -> &Self::Target {
         self.path[0].deref()
-    }
-}
-
-impl<const DEPTH: usize> Default for PoseidonBranch<DEPTH> {
-    fn default() -> Self {
-        let path = iter::repeat(PoseidonLevel::default())
-            .take(DEPTH + 1)
-            .collect();
-
-        Self { path }
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/tree/branch.rs
```diff
@@ -34,6 +31,7 @@ use rkyv::{Archive, Deserialize, Serialize};
 )]
 pub struct PoseidonLevel {
     pub(crate) level: [BlsScalar; dusk_hades::WIDTH],
+    // TODO figure out if we can make this smaller
```
vlopes11: For the level its definitely possible but maybe not worthy since it will be stored in a cpu register anyway.

If `usize`, then it can be `u32` if the target architecture uses that

src/tree/branch.rs
```diff
@@ -79,7 +77,8 @@ impl AsRef<[BlsScalar]> for PoseidonLevel {
     archive_attr(derive(CheckBytes))
 )]
 pub struct PoseidonBranch<const DEPTH: usize> {
-    pub(crate) path: Vec<PoseidonLevel>,
+    path: [PoseidonLevel; DEPTH],
```
vlopes11: it was like this in the past, but then we had some problems in canonical regarding stack allocation. the tree serialization allocated so many branches that often we reached the OS stack limit.

do we have some assurance this won't be the case now?

src/tree/branch.rs
```diff
@@ -103,15 +108,16 @@ impl<const DEPTH: usize> Deref for PoseidonBranch<DEPTH> {
...
-impl<const DEPTH: usize> Default for PoseidonBranch<DEPTH> {
-    fn default() -> Self {
-        let path = iter::repeat(PoseidonLevel::default())
-            .take(DEPTH + 1)
-            .collect();
-
-        Self { path }
-    }
-}
+// // Having a 'Default' branch probably makes no sense
```
vlopes11: Better to remove than comment. `Default` is used mostly for test purposes

src/tree/branch.rs
ureeves: Ah, the classic conundrum of space vs. speed. The comment shouldn't really have been kept here, but it seemed strange to me that a number that is always between 0 and 5 is represented using a `u64`.

src/tree/branch.rs
ureeves: We are indeed not using `canonical` for FFI anymore. If this is too much to be on the stack we can always `Box` the branch.

src/tree/branch.rs
ureeves: Agreed

src/tree/branch.rs
vlopes11: That's true, but then aren't we leaking potential problems? I'm not sure if there is a clear gain to stack allocate these, but I won't block on that since it is highly dependent on the serialization library

vlopes11 (approved):
looks good overall; optional nit

src/tree/branch.rs
```diff
@@ -88,10 +86,16 @@ impl<const DEPTH: usize> PoseidonBranch<DEPTH> {
...
+        &self.root
+    }
+
+    /// Returns a mutable reference to the level with the given `index` in the
+    /// path.
+    ///
+    /// NOTE: This should **never** be exposed to the consumer of the crate.
+    #[allow(dead_code)]
+    pub(crate) fn level_mut(&mut self, index: usize) -> &mut PoseidonLevel {
+        &mut self.path[index]
```
ZER0: At first I was a bit puzzled by the comment and the `#[allow]`, then I noticed this was used in a *unit test* but on a different module.

I would revert the change on the visibility as it was before. I'm not sure why it was changed.
Having  the `path` field directly `pub(crate)` feels like a better ergonomic: beside the extra method only used in the tests, there is no dead code and an extra comment to explain the additional method.

ZER0 (changes requested):

src/tree/branch.rs
ureeves: If I get this right, placing it on the heap was done to skirt around a problem with canonical where it would blow the stack - I guess on deserialization. Since we're not using canonical anymore - and we are removing the feature entirely - we don't need to support this anymore.
`rkyv` our new serialization library - does zero copy deserialization, so I don't expect us blowing the stack like before. In any case it's a problem that can be solved again quickly.

src/tree/branch.rs
ureeves: I think both things are pretty unseemly, but I don't have a strong opinion of one vs. the other. I'll change it back.

ureeves: Done

ZER0 (approved):
LGTM

PR #207, opened 2023-03-27
Add benchmark for sponge gadget

### LINKED ISSUE #206: Add benchmarks for poseidon sponge
#### Summary
Add benchmarks for the sponge gadget proof creation and verification.

Same structure as the merkle opening benchmarks.

#### Additional context
N/A

moCello:
Resolves #206

### DIFF
--- a/benches/sponge.rs
+++ b/benches/sponge.rs
@@ -1,0 +1,96 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use dusk_hades::WIDTH;
+use dusk_plonk::prelude::*;
+use rand::rngs::StdRng;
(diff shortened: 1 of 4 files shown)


ureeves (approved):
LGTM but... could you post the results? 😊

moCello:
> LGTM but... could you post the results?

Sure! I thought the benchmarks are more of a preemptive measurement so we have a way to measure future changes by running them locally before and after the change.

Where would you like to have the results posted? Is it something that you think should be in the README?

benches/sponge.rs
```diff
@@ -0,0 +1,103 @@
...
+    let circuit = SpongeCircuit::new(message);
+
+    // Benchmark sponge native
+    c.bench_function("sponge native", |b| {
+        b.iter(|| {
+            dusk_poseidon::sponge::hash(black_box(&circuit.message));
+        })
+    });
+
+    // Benchmark circuit compilation
```
xevisalle, line 71: I don't think it is useful to benchmark the compilation, plus, it'll take more time to run the benchmarks.

xevisalle (approved):
LGTM with a nit commented in the code.

benches/sponge.rs
moCello: opened issue #209 for this

PR #208, opened 2023-03-28
Add benchmarks for poseidon cipher

### LINKED ISSUE #198: Add benchmarks for PoseidonCipher ZK and non-ZK
#### Summary
We need benchmarks in order to track performance throughout the development.

In this issue we will add the benchmark to measure the native poseidon cipher and the proof creation and verification of the poseidon cipher circuit.

#### Additional context
First mentioned in #81

moCello:
Resolves #198

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Benchmark for running sponge gadged on 5 BlsScalar (one permutation) [#206]
+
 ## [0.28.1] - 2023-01-18
 
 ### Added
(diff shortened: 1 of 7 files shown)


ureeves (approved):
LGTM but... could you post the results? :blush:

xevisalle (approved):
LGTM but I would merge the encrypt and decrypt benchmarks into a single file. Besides, I'd remove the benchmark of the circuit compilation.

moCello:
> LGTM but I would merge the encrypt and decrypt benchmarks into a single file. Besides, I'd remove the benchmark of the circuit compilation.

I get your point with the compilation benchmarks, but regarding the different files for decryption and encryption: the benefit for having them in different files is that one can run them separately with:
```
cargo bench --bench cipher_encrypt
```
or
```
cargo bench --bench cipher_decrypt
```

PR #220, opened 2023-06-07
Add sponge implementation for const generic input len

moCello:
This is needed for halving the gate count of the merkle opening in half


moCello:
closing this PR due to github hickups

PR #221, opened 2023-06-07
Add sponge for merkle hashing over fixed input len

### LINKED ISSUE #215: Add sponge framework over fixed size input
#### Summary
Add a sponge fuction that is generic over the input length and therefore doesn't need to add a padding at the end of the input, saving a full computation when the input length is dividable by the rate.

This should be the default sponge function, as the sponge function with variable input length is expected to be used less.

#### Additional context
See [this paper](https://hackmd.io/bHgsH6mMStCVibM_wYvb2w) for more info

moCello:
Resolves: #215


xevisalle (changes requested):
To me, having to call `sponge::MERKLE::hash(...)` each time I want to hash something, even if it isn't part of the Merkle tree, adds a lot of confusion to the code. I strongly suggest to change this.

moCello:
> To me, having to call sponge::MERKLE::hash(...) each time I want to hash something, even if it isn't part of the Merkle tree, adds a lot of confusion to the code. I strongly suggest to change this.

the normal sponge functions are still used for arbitrary hashing. So when you want to get the poseidon hash for your own type, you need to use the normal `sponge::hash` and `sponge::gadget`. The specific `sponge::merkle::hash` and `sponge::merkle::gadget` functions from this PR are *only* used for the hashing of the levels in a merkle tree. They basically only for internals (i.e. you would only use them when you want to make your own merkle tree)

xevisalle (approved):

PR #230, opened 2023-06-28
Add rkyv features again

moCello:
They were removed when the merkle tree got removed but they are still needed for the cipher


xevisalle (approved):
LGTM

PR #241, opened 2024-01-16
Restructure crate features

### LINKED ISSUE #184: Restructure crate features
The features of a crate should be named accordingly to the features they enable, and not the names or technicalities of its dependencies. I propose a restructuring of the features of this crate.

Currently the features included with this crate are:
```text
default
alloc
rkyv-impl
```

Almost the entire crate sits behind the `alloc` feature with following exceptions:
```text
cipher::PoseidonCipher                                                                                   
perm_uses::two_outputs                                                                                   
sponge::hash                                                                                             
sponge::truncated::hash 
```

I propose that this is terribly confusing, and as such we should rethink the features of the crate.

moCello:
Resolves #184


.github/workflows/dusk_ci.yml
ureeves, line 1: Are the `no_std` tests removed?

Cargo.toml
ureeves, line 1: You add the `cipher` and `zk` features, and remove the `alloc` feature. I think this should probably be added explicitly in the changelog.

src/lib.rs
```diff
@@ -5,12 +5,13 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 #![no_std]
-#![cfg_attr(feature = "alloc", warn(missing_docs), doc = include_str!("../README.md"))]
+#![doc = include_str!("../README.md")]
+#![deny(missing_docs)]
 
-#[cfg(feature = "alloc")]
 extern crate alloc;
```
ureeves: Do we use the `alloc` crate explicitly at all?

ureeves (approved):
Looks good. I would just ask that you add the features changes to the changelog explicitly, and take a look at the crate if the `alloc` crate is used explicitly at all. If not you can also remove `extern crate alloc`.

.github/workflows/dusk_ci.yml
moCello: As far as I understand, since I removed the `default` feature set, all tests are essentially `no_std`. So I removed the `no_std` test.

src/lib.rs
moCello: Actually we don't, good catch! I removed it

PR #242, opened 2024-01-18
Move hades permutation into poseidon lib

### LINKED ISSUE #240: Move hades permutation in this crate
#### Summary
Move the hades-based permutation into this crate.
Poseidon in it's core is a sponge framework that operates on a hades-based permutation. This permutation should also be defined in the same crate.

moCello:
Resolves #240


src/lib.rs
```diff
@@ -18,3 +18,6 @@ pub mod perm_uses;
 
 /// Implementation for the Poseidon Sponge hash function
 pub mod sponge;
+
+/// Implementation of the Poseidon permutation based on the Hades strategy
+pub mod hades;
```
xevisalle, line 23: Is pub necessary?

xevisalle (approved):
LGTM

src/lib.rs
moCello: That is indeed debatable! For this PR I wanted to simply port everything from hades without doing a lot of changes and since the crate was public, I though it makes sense to make the `hades` module in our poseidon library public as well. I opened an issue to only expose two permutation functions (one normal and the gadget) and the hades `WIDTH` #243

moCello: I'll merge this PR and we continue the discussion in #243

PR #254, opened 2024-03-19
Add encryption and decryption

### LINKED ISSUE #236: Allow to encrypt different plaintext sizes
#### Summary
At the moment, PoseidonCipher allows to encrypt messages whose size is always 2 BlsScalars. We need to allow for encryption of custom sizes.

For encrypting Citadel Requests, we are currently using 3 instances of the cipher, which involves needed to store onchain 2 extra nonces that are unnecessary. For encrypting Citadel Licenses, we need two instances, which means 1 extra nonce onchain.

This approach would be useful also for standard Phoenix Notes, if we allow to include the metadata per each note.

moCello:
Resolves #236


ureeves (approved):
Very nice! LGTM!

PR #262, opened 2024-04-17
Change finalize to return the hash directly

### LINKED ISSUE #259: Let `finalize` return the hash directly and not wrapped into a `Result`
#### Summary
Change `finalize` in both `Hash` and `HashGadget` to return the hash directly. We can do this since we start and finish the `Safe` instance in the same method and hence have direct control over the correctness of the io-pattern (which is the only way the method would error), i.e. an error is in fact not possible.

moCello:
Resolves #259


xevisalle (approved):

ureeves (approved):
Nice! LGTM

PR #276, opened 2026-03-27
AI-friendliness pass

moCello:
## Summary

Add Makefile, update CI to use inline `make` targets on `core` runners with `dsherret/rust-toolchain-file@v1`, add a `no-std` bare-metal CI job, update `rustfmt.toml` with import grouping and field shorthand, reformat code, create `AGENTS.md` with full project documentation, and replace `CLAUDE.md` with a symlink to it.


xevisalle (approved):
LGTM

PR #282, opened 2026-06-18
Apply optimizations

### LINKED ISSUE #281: Optimize Poseidon permutation constraints without changing hash outputs
#### Summary
The current Poseidon/Hades252 gadget applies a dense width-5 MDS matrix after every round. In partial rounds, this costs 10 addition gates per round even though only one state element passes through the S-box. Since the crate is used in ZK circuits and consensus-critical hash paths, reducing these constraints without changing outputs would improve proving performance while preserving compatibility.

Implement the standard optimized Poseidon schedule by deriving compressed round constants, a pre-sparse matrix, and sparse partial-round matrices from the existing Hades252 constants and MDS matrix. The implementation must preserve all existing hash outputs, native/gadget equivalence, encryption behavior, and no_std support.

xevisalle:
Closes: #281 

This PR optimizes the internal Poseidon/Hades252 permutation schedule and the PLONK gadget representation while preserving the existing hash function, outputs, and public APIs.

### Changes

  - Adds precomputed optimized constants derived from the existing `assets/arc.bin` and `assets/mds.bin`.
  - Replaces the internal round schedule with an algebraically equivalent optimized schedule using compressed constants, one pre-sparse matrix, and
  sparse MDS matrices for partial rounds.
  - Updates both scalar and PLONK gadget permutation implementations.
  - Further optimizes the PLONK gadget by:
    - folding the first full-round constants directly into the first S-box constraints;
    - representing partial-round sparse MDS state as bounded linear expressions for two rounds at a time;
    - materializing those expressions only when needed by an S-box or later full round.
  - Adds a test-only legacy permutation implementation and verifies it matches the optimized permutation over fixed and deterministic randomized
  states.
  - Adds a negative hash-gadget test proving that an incorrect public digest cannot satisfy the circuit.
  - Adds documentation for the gadget constraint optimization.
  - Updates the changelog.

### Why this remains safe

  - This does not introduce Poseidon2 or new Poseidon parameters.
  - It does not change the field, width, alpha, round counts, round constants, MDS matrix, SAFE domains, serialization, or output encoding.
  - Existing fixed Poseidon hash vectors still pass.
  - The `optimized_matches_legacy_permutation` test compares the optimized implementation directly against the original round-constant/MDS schedule.
  - Gadget outputs are still checked against native outputs by existing proof tests.
  - The circuit remains constrained: S-box inputs are materialized before non-linear gates, linear expressions are constrained when materialized, and
  constants remain fixed in code.
  - Encryption/decryption gadget negative tests still pass.
  - The new `incorrect_hash_output_fails` test checks that a wrong hash output cannot be proven.

### Expected impact

  - Sparse partial-round MDS reduces partial-round linear-layer cost from 10 add gates to 6 add gates per partial round.
  - The first optimization step reduces the approximate raw permutation cost from about 985 gates to about 745 gates.
  - The additional gadget-local expression optimization reduces the measured raw permutation cost further to about 650 gates with `dusk-plonk` 0.23.
  - These counts are performance measurements, not protocol invariants; Poseidon semantics and outputs are unchanged.


HDauven (commented):
LGTM but can we include a deterministic generator or test for optimized_constants.rs? Unless you and @moCello think it's unnecessary.

The comments say OPT_ROUND_CONSTANTS, PRE_SPARSE_MATRIX and SPARSE_MATRICES are derived from `assets/arc.bin` and `assets/mds.bin`. Because this is consensus critical hash code, I'd like the PR to make that derivation reproducible rather than requiring reviewers to trust a large scalar table.

Ideally such a test would:
1. load the existing arc.bin and mds.bin values
2. recompute OPT_ROUND_CONSTANTS, PRE_SPARSE_MATRIX and SPARSE_MATRICES
3. assert byte-for-byte equality with the checked-in constants

The existing optimized_matches_legacy_permutation test is useful, but it does not prove the generated constants are correct.

xevisalle:
> LGTM but can we include a deterministic generator or test for optimized_constants.rs? Unless you and @moCello think it's unnecessary.
> 
> The comments say OPT_ROUND_CONSTANTS, PRE_SPARSE_MATRIX and SPARSE_MATRICES are derived from `assets/arc.bin` and `assets/mds.bin`. Because this is consensus critical hash code, I'd like the PR to make that derivation reproducible rather than requiring reviewers to trust a large scalar table.
> 
> Ideally such a test would:
> 1. load the existing arc.bin and mds.bin values
> 2. recompute OPT_ROUND_CONSTANTS, PRE_SPARSE_MATRIX and SPARSE_MATRICES
> 3. assert byte-for-byte equality with the checked-in constants
> 
> The existing optimized_matches_legacy_permutation test is useful, but it does not prove the generated constants are correct.

Sounds good. Let's see what Moana thinks.

src/hades/optimized_constants.rs
```diff
@@ -0,0 +1,4142 @@
...
+//! Optimized Poseidon constants derived from `assets/arc.bin` and
+//! `assets/mds.bin`.
+//!
+//! These constants encode an algebraically equivalent schedule for the existing
+//! Hades252 permutation. They do not define a new hash function.
+
+use dusk_curves::bls12_381::BlsScalar;
+
+use crate::hades::{PARTIAL_ROUNDS, WIDTH};
+pub(crate) const OPT_ROUND_CONSTANTS: [BlsScalar; 100] = [
```
moCello, line 16: **WARNING:** Settling the open question from the earlier round: document the generator for these constants in `assets/HOWTO.md` before merge — 665 checked-in scalars currently cannot be traced to `assets/arc.bin`/`assets/mds.bin` from anything in the repo.

**Context:** Correctness is already pinned — `optimized_matches_legacy_permutation` recomputes the legacy schedule from the same source tables and the fixed vectors hold — so what is missing is provenance, not verification. `assets/HOWTO.md` covers only `arc.bin` and `mds.bin`; a reviewer cannot see where `OPT_ROUND_CONSTANTS`, `PRE_SPARSE_MATRIX`, and `SPARSE_MATRICES` come from.

**Fix:** Extend `assets/HOWTO.md` with the derivation: the algorithm (constant folding through the inverse partial-round layers, the sparse factorization), the generator used, and how to run it against `arc.bin`/`mds.bin`. A `#[cfg(test)]` derivation asserting byte-equality with the checked-in tables would be even better and is welcome as a follow-up, but is not asked for here.

src/hades/permutation/gadget.rs
```diff
@@ -46,90 +275,152 @@ impl<'a> Safe<Witness, WIDTH> for GadgetPermutation<'a> {
...
+
+        state.copy_from_slice(&result);
+    }
+
+    fn mul_sparse_matrix(
+        &mut self,
+        round: usize,
+        state: &mut [Witness; WIDTH],
+    ) {
+        let mut result = [Composer::ZERO; WIDTH];
```
moCello, line 360: **WARNING:** The `mul_sparse_matrix` and `add_round_constants` trait impls are dead code that would emit constraints if called and look live; the restructure removing them can land as a follow-up.

**Context:** The gadget's `perm` override replaces the trait-default partial rounds with `apply_partial_rounds` (so `apply_partial_round` → `mul_sparse_matrix` is never called) and the initial constant addition with `apply_first_full_round` (so `add_round_constants` is never called). Trait-impl methods always count as used, so no dead-code lint fires, and no test exercises either path.

**Fix:** Restructure the trait so each side owns its round blocks and `perm` exists once:

```rust
pub(crate) trait Hades<T> {
    // building blocks, one impl per representation (unchanged)
    fn quintic_s_box(&mut self, value: &mut T, post_constant: BlsScalar);
    fn mul_matrix(&mut self, state: &mut [T; WIDTH]);
    fn mul_pre_sparse_matrix(&mut self, state: &mut [T; WIDTH]);

    // required hooks — scalar and gadget each implement their own
    fn apply_first_full_round(&mut self, offset: &mut usize, state: &mut [T; WIDTH]);
    fn apply_partial_rounds(&mut self, offset: &mut usize, state: &mut [T; WIDTH]);

    // shared defaults
    fn apply_full_round(/* unchanged */) { /* unchanged */ }
    fn perm(&mut self, state: &mut [T; WIDTH]) { /* single copy, this override's shape */ }
}
```

`add_round_constants`, `apply_partial_round`, and `mul_sparse_matrix` leave the trait and become methods of the scalar impl; the gadget's existing inherent hooks become its trait-impl methods, and both unused gadget impls disappear along with the duplicated `perm` override. The hooks must stay default-free: a `T`-generic default could only delegate back to `add_round_constants`/`mul_sparse_matrix`, which would pin them in the trait and keep these impls mandatory.

src/hades/permutation/gadget.rs
```diff
@@ -46,90 +275,152 @@ impl<'a> Safe<Witness, WIDTH> for GadgetPermutation<'a> {
...
         }
 
         state.copy_from_slice(&result);
     }
+
+    fn perm(&mut self, state: &mut [Witness; WIDTH]) {
+        state.reverse();
+
+        let mut constants_offset = 0;
+        self.apply_first_full_round(&mut constants_offset, state);
```
moCello, line 399: **NOTE:** The round schedule exists twice — this override and the trait-default `perm` differ only in first- and partial-round handling, so a schedule edit in one can silently miss the other.

**Context:** Divergence would be caught only by the gadget-vs-native proof tests, not by the compiler.

**Fix:** Share one default `perm` in the trait, with the first full round and the partial-round block as required hooks each side implements — the trait sketch is in the comment on `mul_sparse_matrix` above; one restructure resolves both. Pure code motion: both sides' round sequences stay byte-identical, pinned by the equivalence test and the fixed vectors.

src/hades/permutation.rs
```diff
@@ -33,92 +38,121 @@ pub(crate) mod scalar;
...
+
+    /// Coefficient of a sparse matrix.
+    fn sparse(round: usize, row: usize, column: usize) -> BlsScalar {
+        match column {
+            0 => SPARSE_MATRICES[round][row],
+            _ if row == 0 => SPARSE_MATRICES[round][WIDTH + column - 1],
+            _ if row == column => BlsScalar::one(),
+            _ => BlsScalar::zero(),
         }
     }
```
moCello, line 157: **NOTE:** `pre_sparse` and `sparse` are applied transposed relative to `mul_matrix` — the stored tables compensate — but nothing in the names, parameter labels, or docs records this convention.

**Context:** `mul_matrix` computes `result[k] = Σ_j reversed_mds(k, j)·state[j]` while `mul_pre_sparse_matrix` and `mul_sparse_matrix` compute `result[j] = Σ_i pre_sparse(i, j)·state[i]`, in both the scalar and gadget implementations. Regenerating the tables under the usual row-major convention would produce silently transposed values, caught only by the equivalence test.

**Fix:** State the storage orientation in the doc comments of `pre_sparse`/`sparse` and in `optimized_constants.rs`, or swap the index order so all three accessors share one convention.

src/hades/permutation/gadget.rs
```diff
@@ -24,6 +84,175 @@ impl<'a> GadgetPermutation<'a> {
     pub fn new(composer: &'a mut Composer) -> Self {
         Self { composer }
     }
+
+    fn gate_add_expression(
+        &mut self,
+        terms: &[(BlsScalar, Witness)],
+        constant: BlsScalar,
+    ) -> Witness {
+        assert!(terms.len() <= 3);
```
moCello, line 93: **NOTE:** `assert!` panics in library code; the crate convention is to avoid panics outside tests, and both call sites already guarantee `terms.len() <= 3`.

**Context:** The two drop-in alternatives both fall short.
A `debug_assert!` compiles out of every CI run — `make test` passes `--release` — so the invariant would be checked nowhere.
Returning an error instead is not worth the plumbing: the helper sits four layers inside the `Hades` trait machinery (`materialize_expression` → `apply_sparse_matrix_as_expressions` → `perm`), so fallibility would spread into the shared trait signature and the public hash entry points for a condition that is a code bug, not an input — the top-level caller could only unwrap it.

**Fix:** Fold the chaining loop from `materialize_expression` into this helper so it handles any number of terms itself — the length invariant disappears rather than being guarded, and the call sites simplify.

moCello (commented):
COMMENT — the optimized schedule is algebraically sound and well tested against the legacy permutation; nothing blocks outright. On the open constants-reproducibility question: documenting the generator in `assets/HOWTO.md` settles it for this PR — the equivalence test already pins correctness, so provenance is what's missing, and the byte-equality test variant can follow separately if wanted. The other findings can land alongside or as follow-ups.
The single commit "Apply optimizations" carries a +4,795-line consensus-critical change with no body; a subject naming the schedule rewrite and a body summarizing what changed and why outputs are preserved would make the history usable.
Downstream circuits embedding this gadget (poseidon-merkle, phoenix, rusk) will compile to different constraint systems on adoption, so verifier keys derived from them must be regenerated — worth stating in the PR description so the version bump is coordinated.

src/hades/permutation.rs
```diff
@@ -33,92 +38,121 @@ pub(crate) mod scalar;
...
+        state: &mut [T; WIDTH],
+    );
 
     /// Computes `input ^ 5 (mod p)`
     ///
     /// The modulo depends on the input you use. In our case the modulo is done
     /// in respect of the scalar field of the bls12_381 curve
     /// `p = 0x73eda753299d7d483339d80809a1d80553bda402fffe5bfeffffffff00000001`.
-    fn quintic_s_box(&mut self, value: &mut T);
+    fn quintic_s_box(&mut self, value: &mut T, post_constant: BlsScalar);
```
moCello, line 52: **NOTE:** The doc comment is stale — the method now computes `input^5 + post_constant`, not `input^5`.

**Fix:** Update the doc to state that the following round's compressed constant is added after the S-box.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Reduce the Poseidon ZK gadget constraint count while preserving existing hash outputs
```
moCello, line 12: **NOTE:** The changelog entry does not link the tracking issue; the repo's format references issues inline with a link definition at the bottom.

**Fix:** Append `[#281]` to the entry and add `[#281]: https://github.com/dusk-network/Poseidon252/issues/281` to the link block.

assets/derive_optimized_constants.py
```diff
@@ -0,0 +1,326 @@
...
+    return [row[size:] for row in augmented]
+
+
+def optimized_round_constants(round_constants, mds):
+    half_full_rounds = FULL_ROUNDS // 2
+    mds_inverse = matrix_inverse(mds)
+    optimized = round_constants[0][:]
+
+    for round_index in range(1, half_full_rounds):
+        optimized.extend(row_mul(round_constants[round_index], mds_inverse))
```
moCello, line 108: **NOTE:** Constant folding computes `c · M⁻¹` while the permutation applies `M` to a column vector, so the derivation holds only for a symmetric MDS, and a non-symmetric one fails silently.

**Context:** `row_mul` treats the constants as a row vector, giving `c · M⁻¹` = `M⁻ᵀ · c`, while `mul_matrix` applies `M · s`. Those agree only when `M = Mᵀ`. The current `mds.bin` is a Cauchy matrix with `x_i = i`, `y_j = j + 5`, which is symmetric, so the tables derived here are correct and `--check` confirms them against the assets.

The failure mode is the silence, not the current output. On a non-symmetric matrix the script emits wrong tables with no error raised: the `matrix_mul(dense, sparse) != matrix` guard in `sparse_factorization` is expressed in the script's own convention and still passes.

**Fix:** Assert symmetry on the parsed MDS before deriving, and state the requirement in `assets/HOWTO.md` next to the storage-orientation note.

src/hades/permutation/scalar.rs
```diff
@@ -19,6 +19,48 @@ impl ScalarPermutation {
...
+
+    fn apply_partial_round(
+        &mut self,
+        round: usize,
+        constants_offset: &mut usize,
+        state: &mut [BlsScalar; WIDTH],
+    ) {
+        self.quintic_s_box(
+            &mut state[0],
+            crate::hades::OPT_ROUND_CONSTANTS[*constants_offset],
```
moCello, line 41: **NOTE:** `OPT_ROUND_CONSTANTS` and `PARTIAL_ROUNDS` are referenced fully qualified inside the round hooks, where the module convention is a top-level import.

**Context:** Also at lines 119 and 131, in `apply_first_full_round` and `apply_partial_rounds`. `gadget.rs` imports both items at module top and uses bare identifiers throughout.

**Fix:** Import both at the top of `scalar.rs` and drop the `crate::hades::` prefixes.

moCello (approved):
APPROVE — every point raised earlier is resolved on this head: the trait restructure removed the duplicated schedule and dead impls, the constants derivation is now reproducible via `assets/derive_optimized_constants.py --check`, and the equivalence test plus fixed vectors continue to pin the output; CI is green.

Two notes below, neither gating: a silent-failure condition in the generator that only bites on a future parameter change, and an import-style cleanup. Both are small — worth taking in one push rather than separately.

Three further points that ask for nothing on this branch:
- Nothing currently executes `derive_optimized_constants.py`. Running `--check` in CI would catch any future divergence between the checked-in tables and the assets; worth an issue.
- The PR description says the 650-gate measurement was taken "with `dusk-plonk` 0.23"; the repo pins `0.22.0-rc.0` and `docs/gadget-constraints.md` says 0.22 — the description text is the stale one.
- Verifier-key regeneration for downstream circuits is stated in the commit body; adding one line to the PR description would make it visible to release coordination without digging into the commit.

PR #285, opened 2026-08-18
Reject invalid encryption shared secrets

### LINKED ISSUE #284: Reject invalid shared secrets in native encryption APIs
## Summary

The native Poseidon encryption and decryption APIs accept arbitrary `JubJubAffine` shared-secret coordinates. Identity and non-prime-order points should be rejected before they are used as encryption material.

This is preventive library hardening. Current Rusk and Phoenix revisions do not call these native Poseidon encryption APIs; Phoenix derives its note-encryption secret for AES instead, and Phoenix transactions are currently disabled on mainnet.

## Expected behavior

- Native encryption and decryption reject identity, off-curve, pure low-order, and mixed-order shared secrets with `Error::InvalidPoint`.
- Valid prime-order shared secrets retain the existing encryption output and round-trip behavior.
- Regressions distinguish identity, pure low-order, and mixed-order points so a small-order-only check cannot satisfy the tests.
- Gadget APIs clearly document that the calling circuit must constrain point membership and non-identity; this issue must not silently alter circuit layouts or verifier keys.

## Performance

Complete subgroup validation requires a scalar multiplication. The implementation should benchmark and document that cost rather than replacing it with an incomplete small-order check.

HDauven:
## Summary

- reject identity, off-curve, pure low-order, and mixed-order shared secrets in native encryption and decryption
- return the existing `Error::InvalidPoint` before invoking SAFE
- document the membership obligation for gadget callers without changing circuit layouts
- add regressions that distinguish identity, pure low-order, and mixed-order points

## Rationale

Using an identity or non-prime-order DHKE result as encryption material weakens the intended shared-secret boundary. A small-order-only check is insufficient: an on-curve mixed-order point can be non-identity and not itself small-order while still failing subgroup membership.

This is preventive public-library hardening. Current Rusk and Phoenix revisions have no callers of these native Poseidon encryption APIs; Phoenix note encryption feeds its DHKE result to AES instead.

## Compatibility and performance

Valid prime-order inputs reach the unchanged SAFE implementation with identical coordinates and nonce. Gadget constraints are intentionally unchanged because adding point-membership constraints would alter circuit layouts and verifier keys.

Complete subgroup validation adds approximately 75 microseconds per call on the test machine. For two-scalar messages, native encryption measured about 43 to 117 microseconds and decryption about 39 to 116 microseconds. The cheaper on-curve/non-small-order predicate was rejected because it accepts mixed-order points.

## Validation

- `make fmt CHECK=1`
- `make clippy`
- `make test`
- `make check`
- `make doc`
- `make no-std`
- `cargo bench --all-features --no-run`
- mutation controls for unconditional acceptance and the incomplete small-order predicate

Closes #284

### DIFF
--- a/src/encryption.rs
+++ b/src/encryption.rs
@@ -56,7 +56,16 @@ use crate::{Domain, Error};
 /// This function encrypts a given message with a shared secret point on the
 /// jubjub-curve and a bls-scalar nonce using the poseidon hash function.
 ///
-/// The shared secret is expected to be a valid point on the jubjub-curve.
+/// The shared secret must be a non-identity point in the prime-order subgroup
+/// of the jubjub curve.
+/// Invalid shared secrets return [`Error::InvalidPoint`]. This validation costs
+/// one jubjub scalar multiplication per call.
+///
+/// This check does not make DHKE with an unchecked peer point safe. Callers
+/// must validate untrusted peer public points as on-curve, non-identity, and
+/// prime-order before scalar multiplication. Protocol-facing code should not
+/// expose [`Error::InvalidPoint`] separately from authentication or decryption
+/// failures.
 ///
 /// The cipher-text will always yield exactly one element more than the message.
 pub fn encrypt(
@@ -64,6 +73,8 @@ pub fn encrypt(
     shared_secret: &JubJubAffine,
     nonce: &BlsScalar,
 ) -> Result<Vec<BlsScalar>, Error> {
+    validate_shared_secret(shared_secret)?;
+
     Ok(dusk_safe::encrypt(
         ScalarPermutation::new(),
         Domain::Encryption,
@@ -77,7 +88,16 @@ pub fn encrypt(
 /// secret point on the jubjub-curve and a bls-scalar nonce using the poseidon
 /// hash function.
 ///
-/// The shared secret is expected to be a valid point on the jubjub-curve.
+/// The shared secret must be a non-identity point in the prime-order subgroup
+/// of the jubjub curve.
+/// Invalid shared secrets return [`Error::InvalidPoint`]. This validation costs
+/// one jubjub scalar multiplication per call.
+///
+/// This check does not make DHKE with an unchecked peer point safe. Callers
+/// must validate untrusted peer public points as on-curve, non-identity, and
+/// prime-order before scalar multiplication. Protocol-facing code should not
+/// expose [`Error::InvalidPoint`] separately from authentication or decryption
+/// failures.
 ///
 /// The cipher-text will always yield exactly one element more than the message.
 pub fn decrypt(
@@ -85,6 +105,8 @@ pub fn decrypt(
     shared_secret: &JubJubAffine,
     nonce: &BlsScalar,
 ) -> Result<Vec<BlsScalar>, Error> {
+    validate_shared_secret(shared_secret)?;
+
     Ok(dusk_safe::decrypt(
         ScalarPermutation::new(),
         Domain::Encryption,
@@ -92,4 +114,16 @@ pub fn decrypt(
         &[shared_secret.get_u(), shared_secret.get_v()],
         nonce,
     )?)
+}
+
+fn validate_shared_secret(shared_secret: &JubJubAffine) -> Result<(), Error> {
+    if !bool::from(shared_secret.is_on_curve()) {
+        return Err(Error::InvalidPoint);
+    }
+
+    if !bool::from(shared_secret.is_prime_order()) {
+        return Err(Error::InvalidPoint);
+    }
+
+    Ok(())
 }
--- a/tests/encryption.rs
+++ b/tests/encryption.rs
@@ -45,6 +45,44 @@ fn encrypt_decrypt() -> Result<(), Error> {
 }
 
 #[test]
+fn non_prime_order_shared_secrets_fail() {
+    let message = [BlsScalar::from(42)];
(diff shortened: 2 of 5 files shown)


src/encryption.rs
```diff
@@ -93,3 +99,15 @@ pub fn decrypt(
...
+    let is_valid = shared_secret.is_on_curve()
+        & shared_secret.is_torsion_free()
+        & !shared_secret.is_identity();
+
+    if bool::from(is_valid) {
+        Ok(())
+    } else {
+        Err(Error::InvalidPoint)
+    }
+}
```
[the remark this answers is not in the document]
HDauven: Adjusted validation to reject identity and off-curve inputs before running the expensive torsion-free subgroup check. Valid and other subgroup-invalid inputs still receive the complete validation.

tests/encryption.rs
```diff
@@ -44,6 +44,41 @@ fn encrypt_decrypt() -> Result<(), Error> {
...
+    assert!(bool::from(identity.is_identity()));
+    assert!(bool::from(low_order.is_on_curve()));
+    assert!(!bool::from(low_order.is_torsion_free()));
+    assert!(!bool::from(low_order.is_identity()));
+    assert!(bool::from(mixed_order.is_on_curve()));
+    assert!(!bool::from(mixed_order.is_torsion_free()));
+    assert!(!bool::from(mixed_order.is_identity()));
+    assert!(!bool::from(mixed_order.is_small_order()));
+
+    let invalid_secrets = [identity, low_order, mixed_order];
```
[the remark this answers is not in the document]
HDauven: Added an explicit `(0, 0)` off-curve fixture, pinned its premise with `!is_on_curve()`, and require both native APIs to return `InvalidPoint`.

src/encryption.rs
```diff
@@ -93,3 +99,15 @@ pub fn decrypt(
...
+    if bool::from(shared_secret.is_identity() | !shared_secret.is_on_curve()) {
+        return Err(Error::InvalidPoint);
+    }
+
+    if !bool::from(shared_secret.is_torsion_free()) {
+        return Err(Error::InvalidPoint);
+    }
+
+    Ok(())
+}
```
moCello, line 129: **NOTE:** Rejection latency depends on the shared secret — the identity short-circuit returns before the scalar multiplication that every other invalid point pays for.

**Context:** For a peer key of order `k ∈ {2,4,8}`, the shared secret `s·P` is the identity exactly when `k | s`, so that case takes the fast path while every other rejection runs the ~252-bit ladder in `is_torsion_free`. Both return the same `Error::InvalidPoint`, leaving latency as the only distinguisher; three probes at `k = 2, 4, 8` against a static secret separate `s mod 8`. On-curve is safe to short-circuit on — `s·P` is on-curve whenever `P` is, so it reveals only the caller's own input.

**Severity-rationale:** No caller in the tree exposes this timing today, so the premise is conditional and the tier is capped. It is cheap to close now and awkward to revisit once the validation is treated as settled.

**Fix:** Keep the on-curve early return, then use jubjub's own combinator for the rest: `if !bool::from(shared_secret.is_prime_order()) { return Err(Error::InvalidPoint); }`. `is_prime_order` is `is_torsion_free() & !is_identity()` with a bitwise `&`, so both halves always evaluate and the identity case costs the same as any other rejection.

src/encryption.rs
```diff
@@ -56,14 +56,17 @@ use crate::{Domain, Error};
 /// This function encrypts a given message with a shared secret point on the
 /// jubjub-curve and a bls-scalar nonce using the poseidon hash function.
 ///
-/// The shared secret is expected to be a valid point on the jubjub-curve.
+/// The shared secret must be a non-identity point in the prime-order subgroup
+/// of the jubjub curve.
```
moCello, line 60: **NOTE:** The precondition is stated without its consequences — neither that violating it returns `Error::InvalidPoint`, nor that enforcing it costs a JubJub scalar multiplication on every call.

**Context:** `decrypt` carries the same two lines (:83-84) and the same omission. #284 asks for the validation cost to be documented; it currently appears only in the PR description, which callers reading the API never see.

**Fix:** Append one sentence to both doc comments — that a shared secret outside the prime-order subgroup returns `Error::InvalidPoint`, and that the check costs one scalar multiplication per call.

moCello (approved):
APPROVE — Validation is complete (on-curve, non-identity, full subgroup via constant-time scalar multiplication), the test vectors are correct and each one is asserted to have the predicate profile it claims, and circuit layouts are untouched; nothing gates. Two notes inline, both one-line edits in the same file: the order of the two checks, and what the doc comments leave out.

src/encryption.rs
HDauven: Changed the validator to short-circuit only off-curve inputs, then use `is_prime_order()` for every on-curve candidate. Identity and other on-curve invalid points now perform the same subgroup work before branching.

src/encryption.rs
HDauven: Expanded both native API docs with the `InvalidPoint` result, one-scalar-multiplication cost, mandatory peer-point validation before DHKE, and guidance to coalesce protocol-facing failures.

moCello (approved):
APPROVE — Every prior comment is addressed at this head: the validator is on-curve then `is_prime_order()` with no identity fast path, the off-curve case is in the regression test, and both native API docs state the `InvalidPoint` result and its scalar-multiplication cost. Nothing new gates or asks for a change.

PR #293, opened 2026-09-01
Forbid unsafe code

Neotamandua:
This PR adds `#![forbid(unsafe_code)]` at the crate root.

Unsafe Rust is not currently used. This ensures it cannot be introduced anywhere in the crate without causing a compile-time error.


moCello (approved):
APPROVE — the crate contains no `unsafe`, and no derive or macro it uses expands to any, so the attribute compiles clean across every feature combination CI builds. Nothing to raise.

PR #294, opened 2026-09-01
Add public hash compatibility vectors

Neotamandua:
Adds known-answer test vectors for the public Poseidon hash API, covering all domains and multi-output hashing.

This covers and intends to catch unintended compatibility changes to domain separation or hash outputs.


tests/hash_vectors.rs
```diff
@@ -0,0 +1,89 @@
...
+use dusk_poseidon::{Domain, Hash};
+
+#[test]
+fn public_hash_known_answer_vectors() {
+    let input = [
+        BlsScalar::from(1u64),
+        BlsScalar::from(2u64),
+        BlsScalar::from(3u64),
+        BlsScalar::from(4u64),
+    ];
```
moCello, line 22: **CAUTION:** No vector goes past the sponge rate, so the 6-scalar `Domain::Other` shape that produces every note-tree leaf has no frozen value.

**Context:** The rate is 4 (Hades width 5, capacity 1). Every case here absorbs at most 4 and squeezes at most 2, so all five run a single permutation. `Note::hash_inputs` in Phoenix returns 6 scalars, and the transfer contract hashes exactly those under `Domain::Other` to build the note-tree leaf, so a change to that digest moves every note-tree root and invalidates every published opening.

Block-crossing absorb is already pinned by the Hades known-answer test in `src/hades.rs`, at 5, 6, 8 and 10 scalars — but under a zero tag and a hand-built io-pattern, so what stays unfrozen is `Hash`'s own composition above the rate: `io_pattern`, tag and domain for a multi-block input. Block-crossing squeeze is pinned nowhere at all; that test squeezes a single element. The `zk` tests drive `Hash::digest` at 3, 5 and 15 scalars against the gadget rather than against a fixed output.

**Severity-rationale:** This change exists to freeze the public hash outputs, and the shape it leaves free is the one carrying live state; the fix is two more rows in the table already being added.

**Fix:** Add a `Domain::Other` vector over 6 input scalars, and one with `output_len` above 4 so the squeeze crosses a block too — `MultipleOutputCircuit<4, 7>` in tests/hash.rs already runs an output length above the rate, so that call is legal.

tests/hash_vectors.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    for (domain, input, expected) in [
+        (
+            Domain::Merkle2,
+            &input[..2],
```
moCello, line 27: **NOTE:** Nothing records where the expected values came from, or that they are frozen and not to be regenerated.

**Context:** A maintainer whose change turns this red has two readings available — broken compatibility, or stale constants. The file name and test name point at the first; nothing states it. Regenerating the values from the new implementation makes the test green again and removes the protection it adds. The Hades known-answer test in `src/hades.rs` at least records the setup its values were produced under, the zero capacity element and the one-scalar padding.

**Fix:** Add a comment above the table recording how the values were produced and stating that a failure is a compatibility break, not a cue to regenerate.

tests/hash_vectors.rs
```diff
@@ -0,0 +1,89 @@
...
+                0x33, 0x9e, 0x61, 0x6a, 0x65, 0x78, 0xa4, 0xac, 0x99, 0x77,
+                0x97, 0x6b,
+            ],
+        ),
+    ] {
+        assert_eq!(
+            Hash::digest(domain, input)[0].to_bytes(),
+            expected,
+            "{domain:?}"
+        );
```
moCello, line 81: **CAUTION:** `digest_truncated` has no frozen value, and its 250-bit width is implemented twice — once natively, once in the gadget.

**Context:** `finalize_truncated` masks to 250 bits through `TRUNCATION_MASK` (src/hash.rs:167-180); the gadget reaches the same width by a different route, `append_logic_xor::<125>` (src/hash/gadget.rs:95). Nothing asserts the two agree except `tests/hash.rs`, which is `#![cfg(feature = "zk")]` and compares gadget against native over random input — so a change moving both stays green, and under a bare `cargo test` the truncated path has no coverage at all. A frozen native value is what would give that gadget-vs-native comparison something fixed to drift away from.

The width is a compatibility constant on a live path, not a test-only one: `jubjub-schnorr` calls `digest_truncated` for the signature challenge hash (src/signatures.rs:130), for multisig challenges (src/multisig.rs:319-371) and for hedged nonce derivation (src/nonce.rs:42-81), and that crate is a dependency of `rusk-core`, `phoenix-core` and `phoenix-circuits`. Changing the width invalidates every existing signature.

**Fix:** Add a `Hash::digest_truncated` vector for `Domain::Other`. Its output is a `JubJubScalar`, so it will not fit the existing `(Domain, &[BlsScalar], [u8; 32])` table and needs its own assert; generate the value once against the current implementation and freeze it.

tests/hash_vectors.rs
```diff
@@ -0,0 +1,89 @@
...
+            [
+                0x07, 0xba, 0x3d, 0xa0, 0xb3, 0x2e, 0x6e, 0x53, 0x7e, 0x2c,
+                0x24, 0xf3, 0x78, 0xb5, 0x66, 0x89, 0xe0, 0x41, 0xcb, 0x91,
+                0x62, 0xe0, 0xd6, 0x5b, 0xc8, 0xf4, 0xd4, 0xb2, 0xbe, 0x39,
+                0x8d, 0x6c,
+            ],
+        ),
+        (
+            Domain::Encryption,
+            &input[..3],
```
moCello, line 47: **NOTE:** The encryption construction has no frozen output anywhere in the crate — a follow-up, not work for this branch.

**Context:** This entry pins the `Domain::Encryption` separator through the sponge. `encrypt` passes that same domain into `dusk_safe::encrypt`, which builds a different io-pattern entirely — absorb the shared secret, absorb the nonce, squeeze the message length, absorb the message, squeeze once more — so this vector says nothing about what `encrypt` produces. `tests/encryption.rs` round-trips and checks error paths, so a change that moves ciphertexts stays green there, and Phoenix note encryption is on-chain data.

**Fix:** Open a follow-up issue for known-answer vectors over `encrypt`/`decrypt` under the `encryption` feature.

moCello (commented):
COMMENT — the vectors match the current implementation, the `Encryption`/`Other` pair over identical input genuinely separates the two tags, and the file is ungated so it runs in every feature configuration. Nothing here is wrong and nothing gates the merge; the comments below are about what the freeze does not yet reach.

Two of them are worth a push before this lands, and both are rows in the table you are already adding, so they batch into one:

- No vector goes past the sponge rate of 4. That leaves the 6-scalar `Domain::Other` digest unfrozen — the shape the transfer contract hashes into every note-tree leaf, so it is the one whose output can least afford to move silently. Block-crossing squeeze is unfrozen too; the Hades vectors in `src/hades.rs` cross a block on absorb but squeeze a single element.
- `digest_truncated` has no frozen value anywhere outside the `zk` feature, and its 250-bit width exists twice — `TRUNCATION_MASK` natively, `append_logic_xor::<125>` in the gadget — with only a random-input gadget-vs-native test between them. It is on the signature path: `jubjub-schnorr` uses it for challenge hashes, multisig and hedged nonces.

Two more, priority yours:

- Nothing records where the constants came from or that they must not be regenerated. A maintainer whose change turns this red can make it green again by regenerating, which removes exactly the protection the file adds.
- `encrypt`/`decrypt` output is unpinned anywhere in the crate. That one is a follow-up issue rather than work on this branch.

moCello (approved):
APPROVE — every point from the previous round landed and nothing new turned up: the vectors cross the rate on both absorb and squeeze, the truncated value pins the 250-bit width against the untruncated row, and the provenance comment closes the regenerate-to-green path.

Worth knowing, since the values were generated after the round-schedule optimization rather than before it: the freeze is anchored to released behaviour. The Hades vectors in `src/hades.rs` are byte-identical to v0.41.0, and `src/hash.rs` differs from that tag by an import reorder.

One optional point, not a condition of this approval. The three surfaces — the domain table, the truncated value, the 5-element output — sit in one test function, so the first failing assert hides the rest. The sibling files already split at that granularity; `tests/hash.rs` has separate `test_truncated_gadget` and `test_multiple_output`. Splitting would name every surface that moved in a single run. Coverage is identical either way, so this is yours to take or leave.

No findings.

